PR #20, opened 2016-03-03
Add `matchDataLength` custom prop type validator

coopy:
This supports dataAttribute validation in `VictoryBar` and `VictoryArea`.

It can be used like this:

```
  static propTypes = {
    dataAttributes: CustomPropTypes.allOfType([
      CustomPropTypes.matchDataLength,
      PropTypes.oneOfType([
        PropTypes.object,
        PropTypes.arrayOf(PropTypes.object)
      ])
    ]),
  }
```

### DIFF
--- a/test/client/spec/victory-util/prop-types.spec.js
+++ b/test/client/spec/victory-util/prop-types.spec.js
@@ -8,7 +8,8 @@ import {
   domain,
   scale,
   homogeneousArray,
-  deprecated
+  deprecated,
+  matchDataLength
 } from "src/victory-util/prop-types";
 
 describe("prop-types", () => {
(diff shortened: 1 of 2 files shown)


test/client/spec/victory-util/prop-types.spec.js
```diff
@@ -254,4 +255,24 @@ describe("prop-types", () => {
...
+      expect(validate()).to.not.be.an.instanceOf(Error);
+    });
+
+    it("does not return an error when prop has same length as data", () => {
+      expect(validate([{}, {}], [1, 2])).to.not.be.an.instanceOf(Error);
+    });
+
+    it("returns an error when prop doesn't have same length as data", () => {
+      expect(validate([{}], [1, 2])).to.be.an.instanceOf(Error)
+        .and.have.property("message", "Length of data and testProp arrays must match.");
```
divmain: This isn't enforced by ESLint, so it may not matter.  But

```
expect(thing)
  .to.be.instanceOf(Error).and
  .to.be.otherThing;
```

has been the preferred style in the past.

divmain:
LGTM!

test/client/spec/victory-util/prop-types.spec.js
coopy: Thanks! Fixed.

PR #32, opened 2016-03-18
Add local .eslintrc files

coopy:
**DO NOT MERGE YET - DEPENDS ON CHANGE IN ARCHETYPE**
https://github.com/FormidableLabs/builder-victory-component/pull/50

This change adds local `.eslintrc` files that extend the archetype configuration files.

Local .eslintrc files in the tree enables integration with IDE/editor linting plugins, such as [linter-eslint](https://atom.io/packages/linter-eslint).

### DIFF
--- a/test/.eslintrc
+++ b/test/.eslintrc
@@ -1,0 +1,2 @@
+---
+  extends: ../node_modules/builder-victory-component/config/eslint/.eslintrc-test
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,2 @@
+---
+  extends: ./node_modules/builder-victory-component/config/eslint/.eslintrc-source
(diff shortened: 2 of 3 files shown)


test/.eslintrc
```diff
@@ -0,0 +1,2 @@
+---
```
ryan-roemer, line 1: Is this just for IDEs to pick up? (because the builder archetype scripts already handle linting of test code).

coopy, line 1: That's right. The only practical way to configure e.g. [linter-eslint](https://atom.io/packages/linter-eslint#using-eslint) (Atom) or [SublimeLinter](https://github.com/roadhump/SublimeLinter-eslint#settings) is with a local `.eslintrc` file in the project.

ryan-roemer, line 1: :+1:

PR #81, opened 2016-07-07
Add ecologyPlaygroundLoading func to docs

### LINKED ISSUE #23: Handle loading state for Ecology/Component-playground
Thoughts? Is this possible? 

At minimum, could we set a minimum height to the playgrounds, so the jump isn’t so shocking?

ecology-loooading

kylecesmat:
Will load placeholder boxes on initial render and fix https://github.com/FormidableLabs/formidable-landers/issues/23


kylecesmat:
cc @paulathevalley

paulathevalley:
LGTM!

PR #88, opened 2016-07-22
Docs: Move Style component and spellcheck for the props table

paulathevalley:
Resolves https://github.com/FormidableLabs/victory-docs/issues/38
- Move repeated `<Style rules={VictoryTheme />` to `app.js` and delete from every `docs.js` 
- Spellcheck: `@example` should be `@examples` 

/cc @ebrillhart @kylecesmat @bmathews


bmathews:
👍

PR #199, opened 2017-02-15
Chore/replace for loops

DerekMaffett:
Also went ahead and refactored a `propTypes` method to reduce nesting and clarify the control flow of the function.


boygirl (approved):
LGTM

PR #202, opened 2017-02-18
automatically arrayify sort key

DerekMaffett:
Just realized that this makes sense for the most common application of sorting.


DerekMaffett:
Iteration on this concept resulted in the original concept getting scrapped - opening a new branch that reflects the actual intent.

PR #242, opened 2017-05-09
Chore/clean up exports

DerekMaffett:
https://github.com/FormidableLabs/victory/issues/549


boygirl (changes requested):
You should also change the export statements in https://github.com/FormidableLabs/victory-core/blob/master/src/index.js

DerekMaffett:
Ready for re-review. Since we're not using the inner indexes anymore, I removed them as well.

boygirl (approved):
Looks good!

PR #245, opened 2017-05-10
Bug/fix dy calculation

DerekMaffett:
It looks like we had a font-size variable getting factored into the global `text` element offset rather than just the individual `tspan` lines. 

https://github.com/FormidableLabs/victory/issues/577

### DIFF
--- a/src/victory-label/victory-label.js
+++ b/src/victory-label/victory-label.js
@@ -126,13 +126,12 @@ export default class VictoryLabel extends React.Component {
 
   calculateAttributes(props) {
     const style = this.getStyles(props);
-    const fontSize = style[0].fontSize;
     const lineHeight = this.getHeight(props, "lineHeight");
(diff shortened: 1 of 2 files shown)


boygirl:
This changes styling for labels: 

Before:
screen shot 2017-05-10 at 2 44 54 pm

After:
screen shot 2017-05-10 at 2 44 32 pm

DerekMaffett:
I couldn't find a working old version of this - the `fontSize` multiplier was put in about a year ago and it seems like the labels would have been misplaced before that. https://github.com/FormidableLabs/victory/issues/276

I think it was probably intended to only apply to the special adjustments though and not to the `dy` itself. I moved the multiplier into the `getDy` method so the labels are unchanged while custom `dy` values work.

boygirl:
@DerekMaffett much better! Approved!

PR #258, opened 2017-06-26
[WIP] fix "radial" interpolation

chrisbolin:
hey @boygirl, could I get a little bit of help? :) 

I'm trying to address https://github.com/FormidableLabs/victory/issues/558.

I've made two changes for the "radial" interpolation (which seem to work):
- using the `lineRadial` d3 interpolation (as opposed to the nonexistent `curveRadial`
- using `d3Shape.lineRadial()` as opposed to `d3Shape.line()`

However, I hit an error when rendering the line: `this._curve.lineStart is not a function` from https://github.com/d3/d3-shape/blob/master/src/curve/radial.js#L17.

Also, I tried using `d3Shape.line()`, but I got `output.lineStart is not a function` from https://github.com/d3/d3-shape/blob/master/src/line.js#L25.

These errors only occur when using the "radial" interpolation on `VictoryCore.curve`. (However, they are very similar to the error you get when using "bundle" interpolation on VictoryArea: https://github.com/FormidableLabs/victory/issues/558#issuecomment-310666491)

### DIFF
--- a/demo/app.js
+++ b/demo/app.js
@@ -6,6 +6,7 @@ import AnimationDemo from "./victory-animation-demo";
 import LabelDemo from "./victory-label-demo";
 import LegendDemo from "./victory-legend-demo";
 import TooltipDemo from "./victory-tooltip-demo";
+import CurveDemo from "./curve-demo";
 
 class Home extends React.Component {
   render() {
@@ -38,6 +39,7 @@ class App extends React.Component {
     case "/label": Child = LabelDemo; break;
     case "/legend": Child = LegendDemo; break;
     case "/tooltip": Child = TooltipDemo; break;
+    case "/curve": Child = CurveDemo; break;
     default: Child = Home;
     }
     return Child;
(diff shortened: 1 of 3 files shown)


boygirl:
@chrisbolin the bigger question I have is whether supporting radial interpolation really makes sense anymore now that we are supporting polar versions of `VictoryLine` and `VictoryArea`. Those end up using `lineRadial` and then have interpolation options supported on top of that.

boygirl:
I think the right answer for this set of issues is to 
1) drop `radial` as a supported interpolation option for both `VictoryArea` and `VictoryLine`
2) drop `bundle` as a supported interpolation option for `VictoryArea`. See https://github.com/d3/d3-shape#curveBundle

chrisbolin:
sounds good to me! i'll do that

chrisbolin:
fix then will be in `victory-chart`, not this repo. closing

PR #276, opened 2017-07-28
add Data.downsampleSimple and Data.downsampleZoomFriendly

chrisbolin:
Summary below. Details in the code.

`Data.downsampleSimple(data, maxPoints)`

Good (and fast) for general use. Shouldn't be used with Zoom though:

jul-28-2017 12-12-44 - zoom ugly

`Data.downsampleZoomFriendly(data, maxPoints, startingIndex)`

Starting index is the index relative to the _entire unzoomed dataset_. See `downsampleZoomData` in `VictoryZoomContainer` for more details. Looks good while zoomed :) 

jul-28-2017 12-12-57 zoom pretty

### DIFF
--- a/src/victory-util/data.js
+++ b/src/victory-util/data.js
@@ -247,34 +247,15 @@ export default {
   },
 
   /**
-   * Reduces the size of a data array, such that data.length <= maxPoints.
-   * @param {Array} data: an array of data; must be sorted
-   * @param {Number} maxPoints: maximum number of data points to return
-   * @returns {Array} an array of data, a subset of data param
-   */
-  downsampleSimple(data, maxPoints) {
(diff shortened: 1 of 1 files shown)


chrisbolin:
My thoughts...
- Usage with Zoom: i think `Data.downsampleZoomFriendly` is a really good idea for Zoom (hence why I did this)
- Usage with other components: i think `Data.downsampleSimple` is a decent option for other components, but there are better things out there for non-zoom use cases. If we _want_ to invest time, https://github.com/d3fc/d3fc-sample would be a great place to start (I tried add it, but gave up after an hour as I didn't have time to debug the errors). Also see the paper that that lib references: https://skemman.is/bitstream/1946/15343/3/SS_MSthesis.pdf

PR #321, opened 2017-12-15
Immutable support

stefvhuynh:
accessor props, `x` and `y` will still take original `datum` as argument in their callback functions. other props, like `animate`, will take the transformed `datum` which will not be an `immutable` object.

### DIFF
--- a/src/victory-util/data.js
+++ b/src/victory-util/data.js
@@ -115,13 +115,12 @@ export default {
       const evaluatedX = trimmedDatum._x !== undefined ? trimmedDatum._x : accessor.x(datum);
       const evaluatedY = trimmedDatum._y !== undefined ? trimmedDatum._y : accessor.y(datum);
       const y0 = trimmedDatum._y0 !== undefined ? trimmedDatum._y0 : accessor.y0(datum);
-
       const x = evaluatedX !== undefined ? evaluatedX : index;
       const y = evaluatedY !== undefined ? evaluatedY : trimmedDatum;
       const originalValues = y0 === undefined ? { x, y } : { x, y, y0 };
       const privateValues = y0 === undefined ? { _x: x, _y: y } : { _x: x, _y: y, _y0: y0 };
 
-      return dataArr.concat([
+      dataArr.push(
         assign(
           originalValues, trimmedDatum, privateValues,
           // map string data to numeric values, and add names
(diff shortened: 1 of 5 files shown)


stefvhuynh:
@boygirl, initial attempt. still working out kinks.

src/victory-util/data.js
```diff
@@ -63,40 +83,83 @@ export default {
...
     };
-    const data = dataset.map((datum, index) => {
-      const evaluatedX = datum._x !== undefined ? datum._x : accessor.x(datum);
-      const evaluatedY = datum._y !== undefined ? datum._y : accessor.y(datum);
-      const y0 = datum._y0 !== undefined ? datum._y0 : accessor.y0(datum);
+
+    const data = dataset.reduce((dataArr, datum, index) => { // eslint-disable-line complexity
+      const trimmedDatum = Array.isArray(datum) || Immutable.isList(datum)
+        ? datum
+        : this.trimDatum(datum, props);
```
stefvhuynh, line 105: another potential strategy would be to do a "shallow" `toJS` conversion (pull out all the top level properties into an object but don't convert those objects using `toJS`). this way, any props passed in will remain available and we won't need to iterate through the whitelist on every `datum` (for perf reasons).

boygirl:
@stefvhuynh I think this branch is causing some issues with stacked / grouped charts. My guess would be that the `_y0` data variable isn't getting properly set. 

Before:
screen shot 2017-12-18 at 12 03 45 pm

After:
screen shot 2017-12-18 at 12 02 08 pm

You can test your changes with the Lank workflow Roemer set up.  

1) Make sure all your victory directories are siblings
2) cd into `victory-chart` directory
3) run `lank link`
4) run `npm run lank-watch` and leave that running in its own terminal window
5) run `npm run lank-run` 

Changes to `victory-core` will be picked up without you needing to manually rebuild lib.

src/victory-util/data.js
```diff
@@ -2,8 +2,28 @@ import { assign, uniq, range, last, isFunction, property, sortBy } from "lodash"
...
 import Collection from "./collection";
 import Scale from "./scale";
+import Immutable from "./immutable";
 
 export default {
+  datumWhitelist: [
+    // Data coordinates
+    "x", "y", "y0", "_x", "_y", "_y0",
+    // Svg style properties (https://www.w3.org/TR/SVG2/styling.html#PresentationAttributes)
+    "alignment-baseline", "baseline-shift", "clip", "clip-path", "clip-rule", "color",
```
boygirl, line 12: These should all be camelCased

src/victory-util/data.js
boygirl: I think this might be a safer solution.

stefvhuynh: Alright, I'll try it out with that solution.

PR #323, opened 2017-12-18
Immutable support

stefvhuynh:
shallowly converts `data` prop from `immutable` object to js, preserving all props in each data point within `data`. 

corresponding PR in victory chart: https://github.com/FormidableLabs/victory-chart/pull/542

`x` and `y` accessors and other props like `animate` that can take a function with a `datum` argument will receive a shallowly converted `datum` (`datum` will be a js plain object, but the keys will be in their original forms with a few exceptions).

i.e.
```
x={(datum) => {
  // datum is regular js but the keys are the original immutable objects
  return datum["myNestedStructure"].get("myNestedValue") + 10;
}}
animate={
  onEnter: {
    before: (datum) => {
      // datum is regular js
      // datum.someProp will be an immutable object if it was passed in as such
    }
  }
}
```

### DIFF
--- a/src/victory-util/data.js
+++ b/src/victory-util/data.js
@@ -218,11 +218,16 @@ export default {
     if (!isArrayOrIterable) {
       return [];
     }
+
     const key = typeof props[axis] === "undefined" ? axis : props[axis];
     const accessor = Helpers.createAccessor(key);
-    const dataStrings = (props.data)
-        .map((datum) => accessor(datum))
-        .filter((datum) => typeof datum === "string");
+
+    const dataStrings = props.data.reduce((dataArr, datum) => {
+      datum = this.parseDatum(datum);
+      dataArr.push(accessor(datum));
+      return dataArr;
+    }, []).filter((datum) => typeof datum === "string");
+
     // return a unique set of strings
     return dataStrings.reduce((prev, curr) => {
       if (typeof curr !== "undefined" && curr !== null && prev.indexOf(curr) === -1) {
--- a/src/victory-util/data.js
+++ b/src/victory-util/data.js
@@ -87,9 +87,7 @@ export default {
     };
 
     const data = dataset.reduce((dataArr, datum, index) => { // eslint-disable-line complexity
-      datum = Immutable.isImmutable(datum)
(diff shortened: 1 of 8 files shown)


src/victory-util/data.js
```diff
@@ -63,34 +64,45 @@ export default {
...
     };
+
     const accessor = {
       x: Helpers.createAccessor(props.x !== undefined ? props.x : "x"),
       y: Helpers.createAccessor(props.y !== undefined ? props.y : "y"),
       y0: Helpers.createAccessor(props.y0 !== undefined ? props.y0 : "y0")
     };
-    const data = dataset.map((datum, index) => {
+
+    const data = dataset.reduce((dataArr, datum, index) => { // eslint-disable-line complexity
```
stefvhuynh, line 89: using `reduce` here because js arrays and `immutable` `List`s both have `reduce` and we can then return a regular array.

src/victory-util/helpers.js
```diff
@@ -144,7 +145,18 @@ export default {
...
+        if (Array.isArray(key)) {
+          return x.getIn(key);
+        } else if (typeof key === "string") {
+          return x.getIn(key.replace("[", ".").replace("]", "").split("."));
+        } else {
+          return x.get(key);
+        }
+      }
+      return property(key)(x);
+    };
```
stefvhuynh: This is more of a nice-to-have. People could just always specify an accessor function if they are passing in an `immutable` object.

src/victory-util/immutable.js
```diff
@@ -0,0 +1,33 @@
...
+    return !!(x && x[this.IMMUTABLE_MAP]);
+  },
+
+  shallowToJS(x) {
+    return this.isIterable(x) ? x.reduce((prev, curr, key) => {
+      prev[key] = curr;
+      return prev;
+    }, this.isList(x) ? [] : {}) : x;
+  }
+};
```
stefvhuynh, line 36: these methods should work for `immutable@v3.8.x` (current latest version) as well as for `v4.0.x` (in release candidate phase) when it is released.

boygirl (approved):
Works well for me!

PR #346, opened 2018-03-30
Remove lodash omit

chrisbolin:
see https://github.com/FormidableLabs/victory/issues/956 for background and process.

this is all that will need to change in `victory-core`. i won't merge this until I start testing with `victory-chart`

### DIFF
--- a/src/victory-util/helpers.js
+++ b/src/victory-util/helpers.js
@@ -3,6 +3,28 @@ import { defaults, isFunction, property, omit, reduce } from "lodash";
 import Collection from "./collection";
 
 export default {
+  /**
+   * creates an object with some keys excluded
+   * replacement for lodash.omit for performance. does not mimick the entire lodash.omit api
+   * @param {Object} originalObject: created object will be based on this object
+   * @param {Array<String>} keys: an array of keys to omit from the new object
+   * @returns {Object} new object with same properties as originalObject
+   */
+  omit(originalObject, keys = []) {
+    // code based on babel's _objectWithoutProperties
+    const newObject = {};
+    for (const key in originalObject) {
+      if (keys.indexOf(key) >= 0) {
+        continue;
+      }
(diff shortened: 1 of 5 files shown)


src/victory-util/helpers.js
```diff
@@ -1,8 +1,30 @@
...
 
 export default {
+  /**
+   * creates an object with some keys excluded
+   * replacement for lodash.omit for performance. does not mimick the entire lodash.omit api
+   * @param {Object} originalObject: created object will be based on this object
+   * @param {Array<String>} keys: an array of keys to omit from the new object
+   * @returns {Object} new object with same properties as originalObject
+   */
+  omit(originalObject, keys = []) {
```
ryan-roemer: Can you refactor this to **not** be part of the `default` export? 

We've got tickets already to unwind shoving everything in this mega object.

Then just switch to `import { omit } from "../victory-util/helpers";` and we're off!

src/victory-util/helpers.js
```diff
@@ -148,7 +170,7 @@ export default {
 
   modifyProps(props, fallbackProps, role) {
     const theme = props.theme && props.theme[role] ? props.theme[role] : {};
-    const themeProps = omit(theme, ["style"]);
+    const themeProps = this.omit(theme, ["style"]);
```
ryan-roemer: Can just revert this change when `omit` is a named export in this same file.

ryan-roemer (changes requested):
Looking good!

Have a refactor, then ping for re-review.

src/victory-util/helpers.js
chrisbolin: oh nice! yeah i'd prefer that anyway :)

chrisbolin:
@ryan-roemer - i made those changes!

ryan-roemer (approved):
💯

boygirl (approved):
LGTM! Thanks Chris

PR #355, opened 2018-04-11
add fast-deep-equal locally

chrisbolin:
`src/victory-util/fast-deep-equal.js` is exactly equal to https://github.com/FormidableLabs/fast-deep-equal/blob/master/index.js, with the addition on the license and `/* eslint-disable */`. This should finish our deep-equal work in `victory-core`. The final step will be to pull in Formidable's `fast-deep-equal` fork when it's published


boygirl (approved):
Looks good to me!

PR #361, opened 2018-04-16
getScopedEvents: replace deep merge with shallow assign

chrisbolin:
`getScopedEvents`'s `parseEventReturn` performed a deep merge, which scales `n^2` for the data structures used. an `assign` will do.

### DIFF
--- a/src/victory-util/events.js
+++ b/src/victory-util/events.js
@@ -148,7 +148,7 @@ export default {
     const parseEventReturn = (eventReturn, eventKey) => {
       return Array.isArray(eventReturn) ?
         eventReturn.reduce((memo, props) => {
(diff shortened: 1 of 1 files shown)


boygirl (approved):
👍 The only situations I can think of that would be problematic with this change are rather contrived, and can be avoided by users keeping track of the changes they need to cache between events themselves.

chrisbolin:
should note the performance change:

the response time between making a selection with `VictorySelectionContainer` with 2,500 points and the selection being rendered improved by **21 times**.

chrisbolin:
addresses some of the issues in https://github.com/FormidableLabs/victory/issues/873

